PR #2, opened 2020-11-25
Installer Update

StefanHaboeck:
Add gcc and build-essential for objdump installation
Add aha installation for Report Manager

### DIFF
--- a/installer.sh
+++ b/installer.sh
@@ -116,17 +116,18 @@ fi
 
 # aha for html generation
 echo -e "\\n""$ORANGE""$BOLD""Downloading aha""$NC"
-if ! [[ -f "aha-master" ]] ; then
-  cd ./external || exit 1
+if ! [[ -f "external/aha-master" ]] ; then
   apt-get install make
   apt-get install unzip
-  wget https://github.com/theZiz/aha/archive/master.zip
-  unzip ./aha-master.zip
-  cd ./aha-master || exit 1
+  wget https://github.com/theZiz/aha/archive/master.zip -O external/aha-master.zip
+  unzip ./external/aha-master.zip -d ./external
+  rm external/aha-master.zip
+  cd ./external/aha-master || exit 1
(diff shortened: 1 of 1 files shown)


installer.sh
```diff
@@ -111,3 +113,20 @@ if ! [[ -f "external/objdump" ]] ; then
 else
   echo -e "$ORANGE""objdump is already downloaded and compiled""$NC"
 fi
+
+# aha for html generation
+echo -e "\\n""$ORANGE""$BOLD""Downloading aha""$NC"
```
p4cx, line 118: Your code won't work at my system. I tried it with some little changes, please fix it :) 

```
# aha for html generation
echo -e "\\n""$ORANGE""$BOLD""Downloading aha""$NC"
if ! [[ -f "external/aha-master" ]] ; then
  apt-get install make
  apt-get install unzip
  wget https://github.com/theZiz/aha/archive/master.zip -O external/aha-master.zip
  unzip ./external/aha-master.zip -d ./external
  cd ./external/aha-master || exit 1
  echo -e "$ORANGE""$BOLD""Compile aha""$NC"
  make
  cd ../.. || exit 1
  mv "external/aha-master/aha" "external/aha"
  rm -R external/aha-master
else
  echo -e "$ORANGE""aha is already downloaded and compiled""$NC"
fi
```

I don't know, if `aha` needs stuff from the zip archive. I deleted the unzipped archive to keep the external directory clean. If it won't work, change it.

p4cx (changes requested):
Please fix issue, which is addressed in the comment in the code. Thank you.

installer.sh
StefanHaboeck: Thank you, the changes are fine. 
Furthermore I have added rm ./external/aha-master.zip to remove the zip file. Now everything should be clean.

p4cx (approved):

PR #9, opened 2020-12-02
Show CVSS2&3 base values in emba version vulnerability checker

p4cx:
Run installer to get new data to generate new database and have fun with this new feature :)

image


m-1-k-3 (approved):
Tested on different firmware images and it looks good to me

PR #26, opened 2021-01-14
UX: Change the install prompt choices to match other installers and common package managers

nightlark:
Changes the install prompt text from `(y/Y)` to `(y/N)`. Default behavior by just hitting enter seems to be not installing.

### DIFF
--- a/installer.sh
+++ b/installer.sh
@@ -134,7 +134,7 @@ print_tool_info "qemu-user-static" 0 "qemu-mips-static"
 print_tool_info "binwalk" 0
 
 echo -e "\\n""$MAGENTA""$BOLD""Do you want to install/update these applications?""$NC"
-read -p "(y/Y)" -r ANSWER
+read -p "(y/N)" -r ANSWER
 case ${ANSWER:0:1} in
   y|Y )
     echo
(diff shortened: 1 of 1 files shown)


p4cx:
To install, you have to enter `y` or `Y` and then press enter. Without entering a character it acts like, if there was a wrong input (not `y` or `Y`).

nightlark:
This is a UX bug then -- when you go to install a package with most package managers (or installers) they give you a prompt with a variant of `y/n` where the capitalized letter is the default behavior if you don't type anything.

nightlark:
e.g. with apt:
```shell
> sudo apt-get install binutils
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following additional packages will be installed:
  binutils-common binutils-x86-64-linux-gnu libbinutils
Suggested packages:
  binutils-doc
The following packages will be upgraded:
  binutils binutils-common binutils-x86-64-linux-gnu libbinutils
4 upgraded, 0 newly installed, 0 to remove and 173 not upgraded.
Need to get 2527 kB of archives.
After this operation, 0 B of additional disk space will be used.
Do you want to continue? [Y/n]
```

p4cx:
Thanks for your detailed explanation. I didn't think about it at all and didn't notice it. You're right! Thank you for your contribution to emba.

p4cx (approved):

PR #27, opened 2021-01-14
Bugfixes

p4cx:
Fix -m and -s and some other small bugs.

now `-m s09 s15` or `-m S10` - both fine
`-s`now has no point before path and should be everywhere (like in checksec module)


m-1-k-3 (approved):
looks good to me. Thanks for the fixes

PR #44, opened 2021-01-28
Set permissions right

p4cx:
We don't know what happened here, but we missed some +x flags and now we have them all ;)


m-1-k-3 (approved):
Very interesting. Thanks to https://github.com/tthtlc for reporting

PR #77, opened 2021-03-19
Update installer.sh

firmianay:
Seems no need to ask again

### DIFF
--- a/installer.sh
+++ b/installer.sh
@@ -491,28 +491,16 @@ case ${ANSWER:0:1} in
     systemctl daemon-reload
     systemctl start mongod
     systemctl enable mongod
-    
-    if [[ "$FORCE" -eq 0 ]] ; then
-      echo -e "\\n""$MAGENTA""$BOLD""Do you want to download and update the cve-search database?""$NC"
-      read -p "(y/N)" -r ANSWER
-    else
-      echo -e "\\n""$MAGENTA""$BOLD""The cve-search database will be downloaded and updated!""$NC"
-      ANSWER=("y")
-    fi
-    case ${ANSWER:0:1} in
-      y|Y )
         
-        git clone https://github.com/cve-search/cve-search.git external/cve-search
-        cd ./external/cve-search/ || exit 1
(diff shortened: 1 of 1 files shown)


p4cx:
In my opinion, this is as completely correct as it was. The step with the extra query is filling the database and that takes a good while. The installer in its normal form, without the `-f` or so switches, is mainly intended for development. As a pure user you should simply use `-f` or` -d` for a Docker setup and not manually go through the installer.
@m-1-k-3 , what do you think of that?

m-1-k-3:
Thank you @firmianay to check our installer. Probably @firmianay will find something more that we can check on Monday? ;)

m-1-k-3:
@firmianay I have checked it out now and I can see your point. The cve-database should be installed in default as it is a main element of a lot of areas of emba. But, as we got the feedback to be more granular in the possibilities a user has we have implemented the current behavior. As the user gets the help output if he does not pass arguments to the installer it is fine for me.
The only thing I think we should change:
Move the default behavior to the first line of the help output and give a note that this should be the default usage for most user. Probably we should remove the “Force” and change it to “Install all dependencies – typical initial installation” or something like this. @p4cx are you fine with this?

p4cx:
> The only thing I think we should change:
> Move the default behavior to the first line of the help output and give a note that this should be the default usage for most user. Probably we should remove the “Force” and change it to “Install all dependencies – typical initial installation” or something like this. 

That sounds fine. That's how we do it.

firmianay:
Thanks for reply. Providing a "Force" option is of course a good choice, I just feel that some dependencies are not clear, some should be a mandatory component, not optional. Another example is 'git', used in the front process, but installed later again.

m-1-k-3:
Interesting. Probably we need to review it again.

└─$ grep print_tool.*git installer.sh
print_tool_info "git" 1
print_tool_info "git" 1

p4cx:
I will close this PR, the discussion for the review of the installer can now be found here: #85 

Thanks to @firmianay again for your help.

PR #78, opened 2021-03-22
Fix -s switch

p4cx:
Firmware path has to be absolute to allow shortening with `-s`.


m-1-k-3 (changes requested):
I have done a quick test and yes we need to change this.

But on a quick check I have seen that it also brings troubles on firmwares we are going to extract with emba.

For example we now have a path in our logs like the following:
/DIR300B5_FW214WWB01.bin_binwalk_emba/_DIR300B5_FW214WWB01.bin.extracted/squashfs-root/usr/sbin/rt3052esw
wich is somewhere in 
/home/m1k3/git-repos/emba_forked/logs_test1/extractor

So, nobody knows where to look at. 
Probably we should take the path starting from the current emba location?
./logs_test1/extractor/DIR300B5_FW214WWB01.bin_binwalk_emba/_DIR300B5_FW214WWB01.bin.extracted/squashfs-root/usr/sbin/rt3052esw

p4cx:
image

m-1-k-3:
I will do a test tomorrow

m-1-k-3 (changes requested):
it is mostrly working:

└─$ sudo ./emba.sh -s -l ./logs_testing1 -f ./logs_testing/extractor/DIR300B5_FW214WWB01.bin_binwalk_emba/

=> root directory is a sub directory of the used firmware directory.

image

so, the output for the paths should be something like this:

./logs_testing/extractor/DIR300B5_FW214WWB01.bin_binwalk_emba/_DIR300B5_FW214WWB01.bin.extracted/squashfs-root/usr/sbin/wps

p4cx:
Whether you use a firmware binary file or an already extracted filesystem, the short path (`-s`) will be `./[your_log_path]/firmware/[firmware_dir]`.

modules/P05_firmware_bin_extractor.sh
```diff
@@ -208,7 +206,7 @@ binwalking() {
 
   export OUTPUT_DIR_binwalk
   OUTPUT_DIR_binwalk=$(basename "$FIRMWARE_PATH")
-  OUTPUT_DIR_binwalk="$LOG_DIR"/extractor/"$OUTPUT_DIR_binwalk"_binwalk_emba
+  OUTPUT_DIR_binwalk="$FIRMWARE_PATH_CP""/""$OUTPUT_DIR_binwalk"_binwalk_emba
 
   echo
   print_output "[*] Extracting firmware to directory $OUTPUT_DIR_binwalk"
```
m-1-k-3, line 212: we have to adjust also line 235 and 238 otherwise the ipk and deb extractor will fail

emba.sh
```diff
@@ -341,6 +341,9 @@ main()
     PRE_CHECK=0
     print_output "[*] Firmware directory detected." "no_log"
     print_output "    Emba starts with testing the environment." "no_log"
+    print_output "    The provided firmware will be copied to ""$FIRMWARE_PATH_CP" "no_log"
+    cp -R "$FIRMWARE_PATH" "$FIRMWARE_PATH_CP""/""$(basename "$FIRMWARE_PATH")"
+    FIRMWARE_PATH="$FIRMWARE_PATH_CP""/""$(basename "$FIRMWARE_PATH")"
```
m-1-k-3, line 346: I think we are having an issue with our OUTPUT_DIR which is only feeded in module P05 and we are using it in S09. We need to set it up here to ensure it is setup correctly. Include as easy fix 
OUTPUT_DIR="$FIRMWARE_PATH_CP"

m-1-k-3 (changes requested):
I think after the small changes we are ready to go

m-1-k-3 (approved):

m-1-k-3:
tests are looking good. Going to merge it now

PR #91, opened 2021-04-09
remove BAP code

firmianay:
cwe_checker recently changed the default analysis backend from BAP to the newer Ghidra backend


p4cx (approved):
Thank you for your contribution :)

PR #108, opened 2021-05-14
Fix deep search

p4cx:
- Thread safe
- Faster
- Aurel fw image bug (some overflow in bash variables) seems to be gone (please!)

image


m-1-k-3 (approved):
great improvement on this module

PR #136, opened 2021-07-08
Fix sed again

p4cx:
Add some minor bug fixes (like file path in webreport to exploit files)
Web report is now more stable (less regexes and also only replacement in explicit lines).
Creation of combined yara file (import of all available yara modules in external) now in log/tmp directory.
"$LOG_FILE" is available in each module, a get method isn't necessary.


m-1-k-3 (approved):
looks good

PR #160, opened 2021-09-13
Webui updates and sorting msf db

p4cx:
CWE are now linked too and you will also find detailed logs inside the CWE-part of the report. Output is also now sorted with threading activated.
Metasploit database is now sorted and should be the same on the dev machines. 
Added new favicon to web report.


m-1-k-3:
looks good to me
image

To activate it use -c and -W

m-1-k-3 (approved):

PR #176, opened 2021-11-08
Remove broken code

### LINKED ISSUE #174: Files not found by 50_authentication_check
From the logfile, I conclude that certain files in /etc where not found by the tool. Am I correct?
These files are present in the tested firmware image!

```
[+] Check users, groups and authentication
=================================================================

==> Users with UID zero (0)
-----------------------------------------------------------------
[*] Searching accounts with UID 0
[-] /etc/passwd not available

==> Non-unique accounts
-----------------------------------------------------------------
[*] Searching non-unique accounts
[-] /etc/passwd not available

==> Unique group IDs
-----------------------------------------------------------------
[*] Searching non-unique group ID's
[-] /etc/group not available

==> Unique group name
-----------------------------------------------------------------
[*] Searching non-unique group names
[-] /etc/group not available

==> Query user accounts
-----------------------------------------------------------------
[*] Reading system users
[-] /etc/passwd not available

==> Query NIS and NIS+ authentication support
-----------------------------------------------------------------
[*] Check nsswitch.conf
[-] /etc/nsswitch.conf not available
```

p4cx:
Fix #174 - there is a check, if the files are valid and it was broken. But we are checking in the modules, if the paths are valid (`if -d` or `-f`). Remove it, now it works :)

### DIFF
--- a/config/msf_cve-db.txt
+++ b/config/msf_cve-db.txt
@@ -147,6 +147,9 @@
 /usr/share/metasploit-framework/modules/auxiliary/dos/http/rails_action_view.rb:CVE-2013-6414
 /usr/share/metasploit-framework/modules/auxiliary/dos/http/rails_json_float_dos.rb:CVE-2013-4164
 /usr/share/metasploit-framework/modules/auxiliary/dos/http/sonicwall_ssl_format.rb:#CVE-,#noCVE?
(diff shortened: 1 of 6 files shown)


m-1-k-3 (approved):
code looks good

PR #188, opened 2021-11-29
Net change

BenediktMKuehne:
attempting to fix install errors with cve_search-redis service


BenediktMKuehne:
Should fix installation of cve_search 
switches only redis-server back to localhost
adds password for redis

m-1-k-3 (approved):
Thanks for fixing ... working smooth on a fresh installation

PR #203, opened 2021-12-23
make dependency-check failing

BenediktMKuehne:


m-1-k-3:
As I'm working on another module I have seen that the fail condition is currently working:

image

Probably the real issue was line 293: DEP_ERROR=0

BenediktMKuehne:
Functionality was/is fine jup.
What this PR changes is the return/exit value from 0 to 1, when using only the dependency check.
Currently you always get 0 on exit even if the check fails.

m-1-k-3:
I am sorry but I don't get it ... probably I misunderstood the issue. could you please explain it in more detail including your testcases.

image

BenediktMKuehne:
When running `sudo ./emba.sh -d` you get this in the case of any failed check.
image
Which doesn't let you determine if it failed or succeeded solely by the return value. (since they are the same in both cases)
image
But for EMBArk we need this check quite frequently, so I changed the exit conditions.
What I changed is that the dep-check fails on dep-errors and dep-exits and returns `1`.
image

m-1-k-3:
ok, thanks for clarification ... got it now. I will check this PR

helpers/helpers_emba_dependency_check.sh
```diff
@@ -394,7 +394,7 @@ dependency_check()
     print_output "$ORANGE""To install all needed dependencies, run '""$NC""sudo ./installer.sh""$ORANGE""'." "no_log"
     print_output "$ORANGE""Learn more about the installation on the emba wiki: ""$NC""https://github.com/e-m-b-a/emba/wiki/installation\\n" "no_log"
 
-    if { [[ $ONLY_DEP -eq 0 ]] && [[ $FORCE -eq 0 ]] ;} || [[ $DEP_EXIT -gt 0 ]] ; then
+    if [[ $ONLY_DEP -eq 1 ]] || [[ $FORCE -eq 0 ]]; then
```
m-1-k-3, line 397: Are you missing the DEP_EXIT now? Shouldn't it be like the following?
if [[ $ONLY_DEP -eq 1 ]] || [[ $FORCE -eq 0 ]] || [[ $DEP_EXIT -gt 0 ]]; then

BenediktMKuehne:
Right don't know why I changed that after

m-1-k-3 (approved):
sorry that it took so long ... thanks for the identification of this bug. Will now merge

PR #252, opened 2022-05-26
Add exception for s99 webreport

p4cx:
Removing bugs from webreport.

Webreport with bugs.

Webreport with less bugs.

No.

* **Other information**:
Some weird grep bs: https://unix.stackexchange.com/a/45194/469803


m-1-k-3 (approved):
Thank you for the fix

PR #300, opened 2022-09-04
Limit check to modules directory

p4cx:

Easy fix.

@n0x08 created the issue #297 because after running EMBA with the log dir set inside the EMBA directory he found out, that the reference links to some other modules were wrong in the webreport.

Why? It is possible to add `[REF]` tags to modules in the output of EMBA and the webreport generation script checks then for the modules (previously in the main directory of EMBA with ignoring of the file ending) if they exists. There it found the old log files instead of the modules and linked to them and therefore created the mess. 

No wrong links anymore if there are old log files in the EMBA directory.


m-1-k-3 (approved):
thx for the fix

PR #939, opened 2023-12-05
Fix parsing of version number from binary version string

mj138:
Bug fix

For a binary version string like "openssl:openssl:1.0.1j", current code returns "openssl" for the version instead of "1.0.1j", causing false positives to be returned.

Updated code returns the last field of the binary version string, which should be the correct version number.

I don't believe so.

* **Other information**:
I did a quick test to verify this works for the openssl example mentioned above, but did not perform any further testing. Also not sure if a similar change might be necessary for BIN_NAME (should it always be the first field? or should it be the second to last field in the case where there are 3+ fields?).

### DIFF
--- a/modules/F20_vul_aggregator.sh
+++ b/modules/F20_vul_aggregator.sh
@@ -555,7 +555,7 @@ check_cve_sources() {
   local CVE_VER_SOURCES_FILE="${3:-}"
 
   local BIN_VERSION_ONLY=""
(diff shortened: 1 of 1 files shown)


m-1-k-3 (approved):
good catch ... I think we should do the same for BIN_NAME as we normally have something like the following:

bin_name:version

But in some cases we have an identifier like the following:

vendor:bin_name_version

Could you also implement this?

PR #942, opened 2023-12-05
Fix parsing of binary name from binary version string

mj138:
Bug fix

Code currently pulls binary name from the first field of the binary version string, like so:

bin_name:version

But in some cases we have an identifier like the following:

vendor:bin_name:version

Updated code pulls binary name from second to last field of the binary version string, which should better handle cases in which there are 3+ fields, as described above.

Hopefully not!

* **Other information**:


m-1-k-3 (approved):
thx for the fix

PR #1376, opened 2024-11-14
Added "apt install linux-modules-extra" package for proper installation of ubi and nandsim modules

0xr3act0r:
The linux-modules-extra package must be installed to use the ubi and nandsim modules. Otherwise, nandsim and ubi modules cannot be loaded with modprobe during the dependency check phase. (Ubuntu 22.04)


m-1-k-3:
Thank you very much for this valuable fix.
Will it also work if we move this fix to the ubuntu installer module over here? https://github.com/e-m-b-a/emba/blob/3cfb1e6a46e7554e12049c88e010190715711486/installer/ID1_ubuntu_os.sh#L27

0xr3act0r:
> Thank you very much for this valuable fix. Will it also work if we move this fix to the ubuntu installer module over here?
> 
> https://github.com/e-m-b-a/emba/blob/3cfb1e6a46e7554e12049c88e010190715711486/installer/ID1_ubuntu_os.sh#L27

Of course, it might make more sense. My commit was a bit of a quick fix :-)

m-1-k-3:
Thank you for your updates. Two further points:

* please check the shellcheck results
image
* If we can install it via "print_tool_info" (which will install it afterwards here https://github.com/e-m-b-a/emba/blob/e973ab6c259265eb8d513e28cc798b40fe3011cd/installer/ID1_ubuntu_os.sh#L42) we can use the typical installation mechanism and further checks already available in EMBA. Would this also work?

0xr3act0r:
I hope it is appropriate now, thank you very much for your valuable feedback :)

m-1-k-3 (approved):

m-1-k-3:
Thank you for effort and bringing this fix to EMBA

PR #1415, opened 2024-12-19
Fix missing kernel config analysis because of lKCONFIG_EXTRACTED/KCON…

chconil:
Kernel configuration is not currently not detected, even when the creation is successful. This fixes it.
Feel free to edit the patch


m-1-k-3 (approved):
What a nasty refactoring bug. Thanks for fixing it!

m-1-k-3 (changes requested):
looks as you are using tabs instead of spaces:

image

As we are using spaces could you please change this to use two spaces for a tab?

chconil:
> looks as you are using tabs instead of spaces:
> 
> image
> 
> As we are using spaces could you please change this to use two spaces for a tab?

Weird, my editor should have catched that. Anyway, fixed ! :)

m-1-k-3 (approved):

PR #1615, opened 2025-06-19
Issue1614/feature request  reuse existing SBOM for CVE rechecking without full rescan

Jeff-Rowell:

    - Feature that introduces a new flag `-R` which requires an existing log directory specified by `-l`. Rescans existing SBOM and generates a new VEX JSON report with fresh CVEs

    - Related to: #1614

    - Added `-R` which allows a rescan of the SBOM:
    
      Screenshot 2025-06-18 225228
    
    - Requires a log directory from a previous scan:
    
      Screenshot 2025-06-18 223425

      image

    
    - All modules are skipped besides the F17 module. The feature will backup the existing SBOM+VEX JSON files using the following file naming convention:
      - EMBA_sbom_vex_only.json - Latest version
      - EMBA_sbom_vex_only.previous.json - Previous version
      - EMBA_sbom_vex_only.previous_1.json - 2nd previous version
      - EMBA_sbom_vex_only.previous_2.json - 3rd previous version
      
    - Here is an example of the file backup functionality:
    
      Screenshot 2025-06-18 222517
    
    - SBOM rescans only take about 10 minutes opposed to running a full `default-scan.emba` for 60-90 minutes:
    
      Screenshot 2025-06-18 222559

    - No

* **Other information**:

    - There is no diff functionality between scans n and n-1. I was planning on making a separate diff feature request that would take two SBOM+VEX JSONs or two log directories and compare the changes. A diff feature would not only be useful for comparing vuln diffs upon re-scans, but also for diffing firmware version increments to see what's been fixed or what's been added between versions of a given vendor and model of firmware.


emba
```diff
@@ -368,6 +368,20 @@ main() {
   emba_parameter_parsing "$@"
   set_log_paths
 
+  # Validate RESCAN_SBOM mode requirements
+  if [[ ${RESCAN_SBOM} -eq 1 ]]; then
```
m-1-k-3: We have started quoting of all variable usages a while ago. This is not needed in every single case but it makes further steps simpler as we do not need to check for issues around this.
```
  if [[ "${RESCAN_SBOM}" -eq 1 ]]; then
```

Thanks

emba
```diff
@@ -518,7 +532,7 @@ main() {
 
     # Print additional information about the firmware (-Y, -X, -Z, -N)
     print_firmware_info "${FW_VENDOR}" "${FW_VERSION}" "${FW_DEVICE}" "${FW_NOTES}"
-    if [[ "${KERNEL}" -ne 1 ]] && [[ "${CONTAINER_EXTRACT}" -ne 1 ]] && [[ "${ONLY_DEP}" -eq 0 ]]; then
+    if [[ "${KERNEL}" -ne 1 ]] && [[ "${CONTAINER_EXTRACT}" -ne 1 ]] && [[ "${ONLY_DEP}" -eq 0 ]] && [[ ${RESCAN_SBOM} -eq 0 ]]; then
```
m-1-k-3: see above

m-1-k-3: Please check also the others ...

emba
```diff
@@ -368,6 +368,20 @@ main() {
   emba_parameter_parsing "$@"
   set_log_paths
 
+  # Validate RESCAN_SBOM mode requirements
+  if [[ ${RESCAN_SBOM} -eq 1 ]]; then
+    if [[ -z "${LOG_DIR}" ]] || [[ "${LOG_DIR}" == *"/logs" ]]; then
```
m-1-k-3: I think we can check in this mode for the EMBA SBOM json log file. If it is not available we can exit.

Jeff-Rowell:
> Very nice stuff. Could you check my comments. Afterwards I will do some real testing

I've made the requested updates. Please take a look when you have a chance and let me know what else is needed. Thanks!

m-1-k-3:
Hi @Jeff-Rowell,

 some little issues:
* You are testing in dev mode only, If you remove the -D you are running into the issue that -R is not forwarded to the container - please add it here https://github.com/e-m-b-a/emba/blob/c202aea966cf272a3f00fec406a9108084d532fc/emba#L673
* Additionally, if I try to start it the following way I get an error:
```
└─$ sudo ./emba -l ~/firmware-analysis/emba_logs_sbom -S -R 
[*] Rescanning SBOM using existing log directory /home/m1k3/firmware-analysis/emba_logs_sbom
[*] Tue Jun 24 13:16:23 CEST 2025 - Rescan SBOM mode detected.
    EMBA will reuse existing SBOM for fresh CVE analysis.
[-] ERROR: RESCAN_SBOM mode requires an existing log directory specified with -l parameter
[-] The log directory must be from a previous EMBA scan and contain an SBOM
[-] Tue Jun 24 13:16:29 CEST 2025 - Test ended on Tue Jun 24 13:16:29 CEST 2025 and EMBA failed in docker mode!
[*] Tue Jun 24 13:16:30 CEST 2025 - Final cleanup started.
```

emba
```diff
@@ -368,6 +368,28 @@ main() {
   emba_parameter_parsing "$@"
   set_log_paths
 
+  # Validate RESCAN_SBOM mode requirements
+  if [[ "${RESCAN_SBOM}" -eq 1 ]]; then
+    if [[ -z "${LOG_DIR}" ]] || [[ "${LOG_DIR}" == *"/logs" ]]; then
```
m-1-k-3: I think this check is not fully working as expected. Why not checking for the SBOM file itself?

```
    if ! [[ -f "${SBOM_LOG_PATH}/EMBA_cyclonedx_sbom.json" ]]; then
```

Or, better way would be to define the SBOM path in the defaults and reuse it everywhere

Jeff-Rowell: I fixed the issue where `-R` wasn't being passed to the docker container:
 
 https://github.com/e-m-b-a/emba/blob/9a631b16745e6c8f8df845995187a7cbecec7fdc/emba#L698

Indeed there was an issue with the `|| [[ "${LOG_DIR}" == *"/logs" ]]` check here:

https://github.com/e-m-b-a/emba/blob/05ac3999f72f5be772a0f94d45f33ee106441085/emba#L371-L373

The issue happens when emba runs in the docker container where `LOG_DIR` gets set to `/logs`. Since I was testing in dev mode only with `-D`, I never ran into that issue. After I fixed the issue where `-R` wasn't being passed to the docker container, I was able to fix this by removing the  `|| [[ "${LOG_DIR}" == *"/logs" ]]` check:

https://github.com/e-m-b-a/emba/blob/9a631b16745e6c8f8df845995187a7cbecec7fdc/emba#L371-L390

As for why I'm not only checking for the SBOM file itself. The intent is to give as specific error messages as possible, so I created three different checks for the different scenarios I could think of:

1. User didn't specify `-l` in their command
2. User did specify `-l`, but the directory doesn't exist
3. User did specify `-l`, and the directory exists, but the directory doesn't have an SBOM (maybe they run a profile that doesn't include F15).

When checking just for the SBOM file we'd only be able to provide a broad error message as opposed to the more specific error messages.

Jeff-Rowell: I also added the SBOM path to defaults here:

https://github.com/e-m-b-a/emba/blob/9a631b16745e6c8f8df845995187a7cbecec7fdc/helpers/helpers_emba_defaults.sh#L200-L202

And updated the references here:

https://github.com/e-m-b-a/emba/blob/9a631b16745e6c8f8df845995187a7cbecec7fdc/emba#L383-L385

And here:

https://github.com/e-m-b-a/emba/blob/9a631b16745e6c8f8df845995187a7cbecec7fdc/modules/F17_cve_bin_tool.sh#L48

Thanks for collaborating with my contribution! It's been fun to start to learn the basics of EMBA as opposed to just running it without understanding some of the code!

m-1-k-3:
While testing your PR I was running into some little issues:
* We check for already available log files here and back them up here https://github.com/e-m-b-a/emba/blob/c202aea966cf272a3f00fec406a9108084d532fc/helpers/helpers_emba_print.sh#L65
We need to disable this in the new mode:
```
+  if [[ -f "${LOG_FILE}" && "${RESCAN_SBOM}" -ne 1 ]]; then
```
Same is needed here for the directories:
https://github.com/e-m-b-a/emba/blob/c202aea966cf272a3f00fec406a9108084d532fc/helpers/helpers_emba_print.sh#L807

* The f17 module needs ${LOG_DIR}/tmp but in restart mode this is not available and needs to be created at the beginning of F17:
```
+  if [[ ! -d "${TMP_DIR}" ]]; then
+    mkdir -p "${TMP_DIR}" || true
+  fi
```
* Additionally, I changed the usage from echo to write_log:
```
-      echo "${lCVSS_SEVERITY}" >> "${TMP_DIR}"/SEVERITY_EXPLOITS.tmp
+      write_log "${lCVSS_SEVERITY}" "${TMP_DIR}"/SEVERITY_EXPLOITS.tmp
```

With this in place it looks not too bad but my new json was not parsable:
```
└─$ jq .components[].id /home/m1k3/firmware-analysis/emba_logs_sbom/SBOM/EMBA_cyclonedx_vex_sbom.json     
jq: parse error: Expected separator between values at line 5069, column 347767
```
I need to debug this further

m-1-k-3:
As I did some wrong assumptions my notes and comments are probably totally wrong and you can just ignore them/I will delete them. I will do another check tomorrow.

Jeff-Rowell:
Thanks for the update! I also saw the jq parse errors you pointed out, but not sure what would be causing that. I will also do some more testing.

m-1-k-3:
@Jeff-Rowell ... I think we got it now up and running. Please do some tests and gimme feedback

Jeff-Rowell:
> @Jeff-Rowell ... I think we got it now up and running. Please do some tests and gimme feedback

All looks much better when testing on my end. I ran three tests and confirmed the backup VEX JSON files are all valid:

Screenshot 2025-06-27 093201

Jeff-Rowell:
@m-1-k-3 One thing I think needs to be updated is the `mapfile` call here:

https://github.com/e-m-b-a/emba/blob/5d8f04a765f5f9540d4e0a4cea50c79526aeccfe/modules/F17_cve_bin_tool.sh#L208-L209

After the first run, the `f17_cve_bin_tool/json` directory contains `.tmp.json` files which causes vulns to get double counted because of the find command. I've made the find command more restrictive to not match the `.tmp.json` and it seems to fix this issue:

image

I'll push this shortly.

Jeff-Rowell:
@m-1-k-3 One final consideration for code cleanup is the `lFILE_SUFFIX` variable. I originally had this set to a date timestamp for the backup files. If you're okay with the current `.previous`, `.previous_1` . . .  `.previous_n` naming done here, I can remove this `lFILE_SUFFIX` since its only ever set to empty string.

 https://github.com/e-m-b-a/emba/blob/4695662791393bdc54ad290dbf827a751d4da19f/modules/F17_cve_bin_tool.sh#L904-L921

modules/F17_cve_bin_tool.sh
```diff
@@ -206,7 +206,7 @@ F17_cve_bin_tool() {
   fi
 
   # now we need to build our full vex json
-  mapfile -t lVEX_JSON_ENTRIES_ARR < <(find "${LOG_PATH_MODULE}/json/" -name "*.json")
+  mapfile -t lVEX_JSON_ENTRIES_ARR < <(find "${LOG_PATH_MODULE}/json/" -name "*.json" ! -name "*.tmp.json")
```
m-1-k-3: I think we can just delete the tmp files after using it here. 

https://github.com/e-m-b-a/emba/blob/c202aea966cf272a3f00fec406a9108084d532fc/modules/F17_cve_bin_tool.sh#L847

I dont know why was it removed ...

Jeff-Rowell: > I don't know why was it removed ...

It actually was never removed. That line was added *after* I forked main.

https://github.com/e-m-b-a/emba/blob/bd8560f48686431a422b48107a0dd1d9f5a497e7/modules/F17_cve_bin_tool.sh#L846-L847

Let me sync my fork then I'll do a final cleanup.

m-1-k-3:
Give me a ping after you are fine with the PR. I will also do another final check and then we should be ready to merge

Jeff-Rowell:
> Give me a ping after you are fine with the PR. I will also do another final check and then we should be ready to merge

I ran some more tests after syncing the fork and I'm fine with the PR now. Please take a look and thanks for your help getting this added!

m-1-k-3 (approved):
Looks good to me. I think we are ready to merge now. Thank you very much for your effort and this great new feature.

PR #1944, opened 2026-04-14
fix: ext filesystem mounting script logic

Spenc3rB:

Bug fix. 

The current logic in P14_ext_mounter checks if a file is found (if the length of the string from `find "${lTMP_EXT_MOUNT}" -mindepth 1 -print -quit` is **non-zero**) and returns "[*] No mounted files..." 

The new logic in P14_ext_mounter checks if a file is found (if the length of the string from `find "${lTMP_EXT_MOUNT}" -mindepth 1 -print -quit` is **zero**) and returns "[*] No mounted files..." properly. This allows EMBA to further analyze binaries within mounted directories using modules such as S12_binary_protection.

Reference: https://www.gnu.org/software/bash/manual/bash.html.

### DIFF
--- a/modules/P14_ext_mounter.sh
+++ b/modules/P14_ext_mounter.sh
@@ -61,7 +61,7 @@ ext_extractor() {
     return
   }
   if mount | grep -q ext_mount; then
-    if [[ -n "$(find "${lTMP_EXT_MOUNT}" -mindepth 1 -print -quit)" ]]; then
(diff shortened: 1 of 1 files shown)


m-1-k-3:
Thank you for this fix. Code scanning has now found another issue here https://github.com/e-m-b-a/emba/pull/1944#discussion_r3079808212 Are you interested in fixing this one also?

m-1-k-3 (approved):

PR #2078, opened 2026-08-17
fixing the installation error for 7.x.x kernerls issue#2057

ahmedboufares:
Bug fix

For the ubuntu with 7.x.x kernels the installation is canceled due to a package not found

It just checks if the package ( linux-modules-extra ) is there than use the old line other wise it just skips it.

no

* **Other information**:


m-1-k-3:
Could you please Check the Workflow results? I think shfmt ist failing

m-1-k-3 (approved):

Issue #1852, opened 2026-01-08
Check, verify and finish csv output for all modules


m-1-k-3:
**Is your feature request related to a problem? Please describe.**
* The csv output is inconsistent on module level. Some modules have csv output, some modules are missing csv output.
* The csv output of all modules should be reviewed to ensure we include the relevant data.
* The generated csv files should be reviewd to ensure they are consistent

m-1-k-3:
I will take car of the Lxx modules which should be the last to close this issue

PR #2086, opened 2026-08-24
Check, verify and finish csv output for all modules  #1852

ahmedboufares:
New CSV files to the most significant modules of emba

Some modules don't emmit a CSV file although they ran in the emba scan

no

* **Other information**:


modules/S50_authentication_check.sh
```diff
@@ -111,10 +112,12 @@ search_shadow() {
         fi
         if [[ "${lHTYPE}" == "unknown" ]]; then
           print_output "[+] Found shadow file ""${ORANGE}$(print_path "${lSHADOW_FILE}")${GREEN} with possible hash ${ORANGE}${lHASH}${NC}"
+          write_csv_log "shadow hash" "${lSHADOW_FILE}" "NA" "${lHASH}" "unknown"
```
m-1-k-3, line 115: unknown hashtype vs NA hashtype in line 150.

m-1-k-3 (changes requested):
very good job. One little comment

modules/S50_authentication_check.sh
ahmedboufares: for that part unknown stands for checking it and not finding anything.NA stands for not applicable meaning that we didnt check the type of the hash.

m-1-k-3 (approved):
Thank you

